PR #3, opened 2019-05-23
Update go-tun2socks to v1.14.2

alalamav:
* Updates go-tun2socks to v1.14.2.
* Updates transitive depencies.
* Creates directories for symlinking modules.


bemasc:
:+1: Verified that this gets us the latest API, and builds more cleanly.

PR #5, opened 2019-06-03
Expose a summary of socket activity for Intra

bemasc:
These metrics are chosen to match Intra's current
set of socket metrics (except for retry-related
metrics, since retry is not yet implemented in
this Handler).

Notably, this change renames one of the two tun2socks
packages (renamed to "tunnel"), and exposes bindings for
additional packages.  This appears to be necessary
because gobind can only expose types from packages that
are explicitly exported.

### DIFF
--- a/tunnel/intra/tcp.go
+++ b/tunnel/intra/tcp.go
@@ -33,8 +33,8 @@ type tcpHandler struct {
 
 // Usage summary for each TCP socket, reported when it is closed.
 type TCPSocketSummary struct {
-	Download   int64 // Total bytes downloaded.
-	Upload     int64 // Total bytes uploaded.
+	DownloadBytes   int64 // Total bytes downloaded.
+	UploadBytes     int64 // Total bytes uploaded.
 	Duration   int32 // Duration in seconds.
 	ServerPort int16 // The server port.  All values except 80, 443, and 0 are set to -1.
 	Synack     int32 // TCP handshake latency (ms)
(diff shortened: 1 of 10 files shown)


tunnel/intra/udp.go
```diff
@@ -27,16 +27,29 @@ import (
 	"github.com/eycorsican/go-tun2socks/core"
 )
 
+// Summary of a non-DNS UDP association, reported when it is discarded.
+type UDPSocketSummary struct {
+	Upload   int64 // Amount uploaded (bytes)
```
alalamav: Consider appending the unit to the name (i.e. UploadBytes)

tunnel/intra/tcp.go
```diff
@@ -19,36 +19,81 @@ package intra
...
-	fakedns net.Addr
-	truedns net.Addr
+	fakedns  net.Addr
+	truedns  net.Addr
+	listener TCPListener
+}
+
+// Usage summary for each TCP socket, reported when it is closed.
+type TCPSocketSummary struct {
+	Download   int64 // Total bytes downloaded.
```
alalamav: Consider appending the unit to the name (i.e. UploadBytes)

bemasc: Done

tunnel/intra/udp.go
bemasc: Done

PR #9, opened 2019-06-13
Android: retrieve JNI binaries, don't strip debug symbols

alalamav:
* Passes the `-work` flag to `gomobile bind` in order to get Go's working directory and copy JNI binaries.
* Stops stripping debug symbols in Android builds to enable crash reporting tools to extract them.

### DIFF
--- a/build_android.sh
+++ b/build_android.sh
@@ -15,6 +15,16 @@
 # limitations under the License.
 
 readonly BUILD_DIR=build/android
+readonly LOG_FILE=$(mktemp)
 
 rm -rf $BUILD_DIR
-make clean && make android
+make clean && make android 2>&1 | tee $LOG_FILE
+
+# Parse the Go working directory to copy the unstripped JNI binaries so symbols can be uploaded
+# to crash reporting tools.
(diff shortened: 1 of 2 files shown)


build_android.sh
```diff
@@ -15,6 +15,16 @@
...
-make clean && make android
+make clean && make android 2>&1 | tee $LOG_FILE
+
+# Parse the Go working directory to copy the unstripped JNI binaries so symbols can be uploaded
+# to crash reporting tools.
+readonly GO_WORK_DIR=$(cat $LOG_FILE | grep "WORK=" | cut -f2 -d=)
+if [ ! -z $GO_WORK_DIR ]; then
+  echo "Copying JNI binaries from: $GO_WORK_DIR"
+  readonly JNI_DIR=$BUILD_DIR/jni
+  mkdir -p $JNI_DIR && cp -R $GO_WORK_DIR/android/src/main/jniLibs/ $JNI_DIR/
```
bemasc: Should we delete `$GO_WORK_DIR`?  Otherwise these files will accumulate after every build.

alalamav: Good idea, done in 671e861.

PR #12, opened 2019-08-07
Add an option to always apply splitting to HTTPS

bemasc:


android/intra.go
```diff
@@ -35,12 +35,12 @@ type IntraTunnel interface {
 //
 // Throws an exception if the TUN file descriptor cannot be opened, or if the tunnel fails to
 // connect.
-func ConnectIntraTunnel(fd int, fakedns, udpdns, tcpdns string, listener tunnel.IntraListener) (IntraTunnel, error) {
+func ConnectIntraTunnel(fd int, fakedns, udpdns, tcpdns string, alwaysSplitHTTPS bool, listener tunnel.IntraListener) (IntraTunnel, error) {
```
alalamav: nit: should the name be `alwaysSplitHttps`?

tunnel/intra/direct_split.go
```diff
@@ -0,0 +1,69 @@
...
+	return s.conn.Read(buf)
+}
+
+func (s *splitter) CloseRead() error {
+	return s.conn.CloseRead()
+}
+
+// Write-related functions
+func (s *splitter) Write(b []byte) (int, error) {
+	if s.fresh {
```
alalamav: I think flipping the condition for an early return would make this more readable.

tunnel/intra/direct_split.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"io"
+	"net"
+)
+
+type splitter struct {
+	DuplexConn
+	conn  *net.TCPConn
+	fresh bool
```
alalamav: This variable tracks whether we have read from/written to this connection, right? Consider renaming or adding a comment.

android/intra.go
bemasc: That's what I had originally, but `go lint` demanded this version.

tunnel/intra/direct_split.go
bemasc: Done

tunnel/intra/direct_split.go
bemasc: Done.  (Also this was initialized to false!)

PR #17, opened 2019-08-16
Enable logging on Android

bemasc:


bemasc:
Yes, they show up tagged as "GoLog".  Note that this PR sets the default level to Warn because Info has a lot of logspam (e.g. one message for every socket).  For debugging, we'll have to lower this level manually and rebuild (unless you come up with a better solution).

PR #18, opened 2019-08-16
Revert "Delay RST caused by Network Unreachable"

bemasc:
This reverts commit c13d6d53cfba1c8a8d017bb401eddd48658f5028.

It turns out that go-tun2socks doesn't wait for Handle() to
return before returning success from tcpAcceptFn, so delaying
the return does not have the intended effect.

### DIFF
--- a/tunnel/intra/tcp.go
+++ b/tunnel/intra/tcp.go
@@ -19,8 +19,6 @@ package intra
 import (
 	"io"
 	"net"
-	"os"
-	"syscall"
 	"time"
 
 	"github.com/eycorsican/go-tun2socks/common/log"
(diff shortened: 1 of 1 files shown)


bemasc:
In Intra, I "solved" the problem by fully disabling IPv6, matching your code in outline-client.  A fuller solution would require substantial changes to LWIP (unlikely) or switching to a different TCP implementation (presumably netstack).

I'm not aware of any better options.

ignoramous:
> I'm not aware of any better options.

Is this an okay-ish workaround for the current problem (assuming `tun2socks` doesn't have other problems with IPv6 in general): Before creating a tunnel see if a preset known-address is reachable over IPv6 (much like how `netd` checks on connectivity to `Private DNS`). If it cannot be reached then add only the IPv4 routes.

bemasc:
@ignoramous That might work, but you'll need to perform that check on every network configuration change, wait for the result, and then reconfigure the VPN.  It's messy, and error-prone (e.g. if the new network takes a few seconds to become functional after association).  Things get even more complicated if there are multiple active networks with different scopes.

PR #28, opened 2019-10-08
Make DuplexConn a net.Conn

bemasc:
This is a prerequisite for DoH to use the retrier.

This change also adds a configurable TCP handshake
timeout to retrier, and simplifies the direct split code.

### DIFF
--- a/tunnel/intra/direct_split.go
+++ b/tunnel/intra/direct_split.go
@@ -6,51 +6,38 @@ import (
 )
 
 type splitter struct {
-	DuplexConn
-	conn *net.TCPConn
+	*net.TCPConn
 	used bool // Initially false.  Becomes true after the first write.
 }
 
 // DialWithSplit returns a TCP connection that always splits the initial upstream segment.
 // Like net.Conn, it is intended for two-threaded use, with one thread calling
 // Read and CloseRead, and another calling Write, ReadFrom, and CloseWrite.
-func DialWithSplit(network string, addr *net.TCPAddr) (DuplexConn, error) {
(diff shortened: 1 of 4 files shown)


tunnel/intra/direct_split.go
```diff
@@ -6,51 +6,38 @@ import (
...
 type splitter struct {
-	DuplexConn
-	conn *net.TCPConn
+	*net.TCPConn
 	used bool // Initially false.  Becomes true after the first write.
 }
 
 // DialWithSplit returns a TCP connection that always splits the initial upstream segment.
 // Like net.Conn, it is intended for two-threaded use, with one thread calling
 // Read and CloseRead, and another calling Write, ReadFrom, and CloseWrite.
```
alalamav: Does this comment need to be updated per the interface change?

tunnel/intra/retrier.go
```diff
@@ -275,3 +288,52 @@ func (r *retrier) CloseWrite() error {
...
+
+func (r *retrier) SetWriteDeadline(t time.Time) error {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+	r.writeDeadline = t
+	return r.conn.SetWriteDeadline(t)
+}
+
+func (r *retrier) SetDeadline(t time.Time) error {
+	e1 := r.SetReadDeadline(t)
```
alalamav: Per the comment above, do you need to check whether `r.retryCompleted()` before setting the read deadline?

tunnel/intra/direct_split.go
bemasc: I think this comment is still accurate.

PR #35, opened 2019-10-28
Drop server IPs and sockets that aren't working

bemasc:
With this change, server IPs will be disconfirmed if
they are reachable but not correctly functioning,
and sockets will be closed if they start reporting
errors.

This change also adds significantly more debug logging,
which is not enabled in the default build.

### DIFF
--- a/tunnel/intra/doh/doh.go
+++ b/tunnel/intra/doh/doh.go
@@ -204,10 +204,9 @@ func (t *transport) doQuery(q []byte) (response []byte, server *net.TCPAddr, qer
 		qerr = &queryError{BadQuery, fmt.Errorf("Query length is %d", len(q))}
 		return
 	}
-	id0, id1 := q[0], q[1]
-	id := int(id0)*256 + int(id1)
 	// Zero out the query ID.
-	q[0], q[1] = 0, 0
+	id := binary.BigEndian.Uint16(q)
+	binary.BigEndian.PutUint16(q, 0)
 	req, err := http.NewRequest(http.MethodPost, t.url, bytes.NewBuffer(q))
 	if err != nil {
 		qerr = &queryError{InternalError, err}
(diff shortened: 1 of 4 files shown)


tunnel/intra/doh/doh.go
```diff
@@ -190,34 +196,101 @@ func (e *queryError) Unwrap() error {
...
+	// underlying socket and disconfirm the server IP.  Empirically, sockets often
+	// become unresponsive after a network change, causing timeouts on all requests.
+	defer func() {
+		if qerr == nil {
+			return
+		}
+		log.Infof("%d Query failed: %v", id, qerr)
+		if server != nil {
+			log.Debugf("%d Disconfirming %s", id, server.IP.String())
+			t.ips.Get(hostname).Disconfirm(server.IP)
```
alalamav: In the event of a timeout due to a network change, we would reconfirm the server IP when the query is reissued (and successful), right?

tunnel/intra/doh/doh.go
```diff
@@ -190,34 +196,101 @@ func (e *queryError) Unwrap() error {
...
+// Independent of the query's success or failure, this function also returns the
+// address of the server on a best-effort basis, or nil if the address could not
+// be determined.
+func (t *transport) doQuery(q []byte) (response []byte, server *net.TCPAddr, qerr error) {
 	if len(q) < 2 {
 		qerr = &queryError{BadQuery, fmt.Errorf("Query length is %d", len(q))}
 		return
 	}
 	id0, id1 := q[0], q[1]
+	id := int(id0)*256 + int(id1)
```
alalamav: nit: consider using `id := binary.BigEndian.Uint16(q[:1])` and `binary.BigEndian.PutUint16(q[:1], 0)` to avoid working with single bytes.

bemasc: Good idea.  Done.

tunnel/intra/doh/doh.go
bemasc: Yes, at the end of `doQuery`.

PR #36, opened 2019-10-29
Restructure events for compatibility with Firebase

bemasc:
This change reports DOH events in a way that is compatible
with Firebase Performance Monitoring.

### DIFF
--- a/tunnel/intra/doh/doh.go
+++ b/tunnel/intra/doh/doh.go
@@ -53,16 +53,21 @@ const (
 
 // Summary is a summary of a DNS transaction, reported when it is complete.
 type Summary struct {
-	Latency  float64 // Response (or failure) latency in seconds
-	Query    []byte
-	Response []byte
-	Server   string
-	Status   int
+	Latency    float64 // Response (or failure) latency in seconds
+	Query      []byte
+	Response   []byte
+	Server     string
+	Status     int
+	HTTPStatus int // Zero unless Status is Complete or HTTPError
 }
+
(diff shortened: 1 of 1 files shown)


bemasc:
For reference, here's the corresponding change on the Java side (in the Go-DOH experimental branch): https://github.com/Jigsaw-Code/Intra/commit/c5ecde691a85427dfaa303100c2eff2e400b5ea5

tunnel/intra/doh/doh.go
```diff
@@ -53,16 +53,21 @@ const (
...
+	Latency    float64 // Response (or failure) latency in seconds
+	Query      []byte
+	Response   []byte
+	Server     string
+	Status     int
+	HTTPStatus int // Zero unless Status is Complete or HTTPError
 }
 
+// A Token is an opaque handle used to match responses to queries.
+type Token interface{}
```
alalamav: What fields/methods will the Java interface have to match responses to queries? It may be useful to document and enforce them here. Could we use the query ID?

bemasc: I realized that this is hard to understand in isolation, so I added a link to the corresponding code on the Java side.

I would be concerned about using the query ID because different clients could produce colliding IDs.  Also, IDs in general are a form of manual reference counting, with the potential for a leak if an ID is ever registered and then not cleaned up.  The token-passing design avoids the need for a map from IDs to state, and lets the garbage collector control object lifetimes.

alalamav: I see - the approach sounds reasonable to me.

PR #37, opened 2019-11-20
Update fakeListener to support new Listener interface

dmcardle:
2d8ca51e12d950199239a6afe5f6f45daf27bf62 changed the Listener interface by removing `OnTransaction` and adding `OnQuery` and `OnResponse`, but it looks like the `fakeListener` implementation got left behind.

### DIFF
--- a/tunnel/intra/doh/doh_test.go
+++ b/tunnel/intra/doh/doh_test.go
@@ -286,8 +286,13 @@ type fakeListener struct {
 	summary *Summary
 }
 
-func (l *fakeListener) OnTransaction(s *Summary) {
-	l.summary = s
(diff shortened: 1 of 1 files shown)


tunnel/intra/doh/doh_test.go
```diff
@@ -286,8 +286,13 @@ type fakeListener struct {
 	summary *Summary
 }
 
-func (l *fakeListener) OnTransaction(s *Summary) {
-	l.summary = s
+func (l *fakeListener) OnQuery(url string) Token {
+	var tok Token
```
bemasc: Can this just be `return nil`?

dmcardle: Yep!

PR #38, opened 2019-11-20
Use valid DNS queries in doh_test.go

dmcardle:
I started using net.dns.dnsmessage in doh.go to parse the request, with the intent of adding a parse-transform-serialize function to add padding. But I found the parse step failed because the test queries didn't have a valid questions section!

The issue is that {1, 2, 3, 4, 5} encodes a message that claims to have 5 questions, but actually contains zero! (See [RFC1035](https://tools.ietf.org/html/rfc1035) section 4.1.1.)


dmcardle:
This pull request is now subsumed by #39.

PR #39, opened 2019-11-20
Implement RFC7830 padding for queries

dmcardle:


tunnel/intra/doh/padding.go
```diff
@@ -0,0 +1,112 @@
...
+	// RR's data.
+	var extraPadding = kOptPaddingHeaderLen
+
+	// If we don't already have an OPT RR, we'll need to add its
+	// header as well!
+	if !hasOptRr {
+		extraPadding += kOptRrHeaderLen
+	}
+
+	for (msgLen+extraPadding+padSize)%blockSize != 0 {
```
bemasc: I think this can be `return blockSize-(msgLen+extraPadding+padSize)%blockSize`

tunnel/intra/doh/padding.go
```diff
@@ -0,0 +1,112 @@
...
+
+package doh
+
+import (
+	"golang.org/x/net/dns/dnsmessage"
+)
+
+const (
+	OptResourcePaddingCode = 12
+	OptDefaultPaddingLen   = 400
```
bemasc: Let's stick with the RFC 8467 recommendation of 128.  Also, this is more of a "block size" than a "padding length".

tunnel/intra/doh/padding.go
```diff
@@ -0,0 +1,112 @@
...
+				TTL:   0,
+			},
+			Body: &dnsmessage.OPTResource{
+				Options: []dnsmessage.Option{
+					getPadding(len(rawMsg), false),
+				},
+			},
+		})
+	} else {
+		// Search for a padding Option and delete it.
```
bemasc: If a padding option is already present, I would prefer to return `rawMsg` unmodified, on the theory that the client has already decided its padding policy.

tunnel/intra/doh/doh_test.go
```diff
@@ -153,9 +188,25 @@ func TestRequest(t *testing.T) {
...
 	}
-	// The first two bytes are the ID, so they should be set to zero.
-	if !bytes.Equal([]byte{0, 0, 3, 4, 5}, reqBody) {
-		t.Errorf("Unexpected request body %v", reqBody)
+	if len(reqBody)%OptDefaultPaddingLen != 0 {
+		t.Errorf("reqBody has unexpected length: %d", len(reqBody))
+	}
+	reqBodyPaddedAgain, err := AddEdnsPadding(reqBody)
+	if !bytes.Equal(reqBody, reqBodyPaddedAgain) {
+		t.Errorf("Padding should be idempotent\n%v\n%v", reqBody, reqBodyPaddedAgain)
```
bemasc: This is interesting but I think it should be a new test.

tunnel/intra/doh/padding.go
dmcardle: OK, there are a few more edge cases. Let me know what you think of the new version.

dmcardle:
Argh, so one other thing to think about is that the incoming message is not guaranteed to be fully label-compressed.

(As an aside, is it okay to compress messages that were not already compressed?)

I really should unpack and pack the rawMsg before using its length.

tunnel/intra/doh/padding.go
```diff
@@ -0,0 +1,114 @@
...
+	var extraPadding = kOptPaddingHeaderLen
+
+	// If we don't already have an OPT RR, we'll need to add its
+	// header as well!
+	if !hasOptRr {
+		extraPadding += kOptRrHeaderLen
+	}
+
+	var padSize int = blockSize - (msgLen+extraPadding)%blockSize
+	if padSize < 0 {
```
bemasc: This should be impossible?

tunnel/intra/doh/padding.go
```diff
@@ -0,0 +1,114 @@
...
+		padSize *= -1
+	}
+	if padSize%blockSize == 0 {
```
bemasc: You could just return `padSize%blockSize`.

tunnel/intra/doh/padding.go
```diff
@@ -0,0 +1,114 @@
...
+	// RR's data.
+	var extraPadding = kOptPaddingHeaderLen
+
+	// If we don't already have an OPT RR, we'll need to add its
+	// header as well!
+	if !hasOptRr {
+		extraPadding += kOptRrHeaderLen
+	}
+
+	var padSize int = blockSize - (msgLen+extraPadding)%blockSize
```
bemasc: Style preference: `padSize := ...`

tunnel/intra/doh/padding.go
```diff
@@ -0,0 +1,114 @@
...
+	for _, additional := range msg.Additionals {
+		switch additional.Body.(type) {
+		case *dnsmessage.OPTResource:
+			optRes = additional.Body.(*dnsmessage.OPTResource)
+			break
+		}
+	}
+
+	if optRes != nil {
+		// If the message already contains padding, we will
```
bemasc: Just because the message has an OPT doesn't mean it has padding.  Also, we should probably have tests for queries with no OPT, OPT but no padding, and padding.

tunnel/intra/doh/padding.go
```diff
@@ -0,0 +1,114 @@
...
+		// respect the application's padding.
+		return rawMsg, nil
+	}
+
+	// Build the padding option that we will need. We can't use
+	// the length of |rawMsg|, since its labels may be compressed
```
bemasc: Sounds tricky.  Can we add a test?  You could take the same `dnsmessage.Message`, pack it compressed and uncompressed, verify that the sizes are different, and then verify that the sizes are the same after padding.

I do feel I should point out that compression is unlikely to activate in queries, since it only applies when there are multiple names with a shared suffix, and queries normally contain only a single name.

tunnel/intra/doh/padding.go
dmcardle: Oh duh, yes. I swear it made sense when I wrote it.

tunnel/intra/doh/padding.go
dmcardle: Good catch.

tunnel/intra/doh/padding.go
dmcardle: OK, so the library isn't flexible enough to allow me to choose compressed or not. I've sort of inverted your suggestion with two tests.

Test 1: Take bytes of a compressed query, unpack and repack with compression.  Check that the original and repacked queries have the same length.

Test 2: Take bytes of an *uncompressed* query, unpack and repack with compression. Check that the original query is larger than the repacked query.

These tests are really just verifying that the dnsmessage library does what we expect. We should also test that an uncompressed query comes out padded to the right size.

dmcardle:
@bemasc Don't forget me :)

tunnel/intra/doh/padding.go
bemasc: > OK, so the library isn't flexible enough to allow me to choose compressed or not.

It looks like there might be a way:
https://godoc.org/golang.org/x/net/dns/dnsmessage#Builder.EnableCompression

> These tests are really just verifying that the dnsmessage library does what we expect.

That doesn't sound great.  I think the real goal is to test that this does the right thing when the input is not compressed (or is compressed differently from how `dnsmessage` would do it).

tunnel/intra/doh/padding.go
```diff
@@ -0,0 +1,111 @@
...
+	if err := msg.Unpack(rawMsg); err != nil {
+		return nil, err
+	}
+
+	// Search for OPT resource and save |optRes| pointer if possible.
+	var optRes *dnsmessage.OPTResource = nil
+	for _, additional := range msg.Additionals {
+		switch additional.Body.(type) {
+		case *dnsmessage.OPTResource:
+			optRes = additional.Body.(*dnsmessage.OPTResource)
```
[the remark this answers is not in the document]
dmcardle: So I'm iterating over a list of `Resource` values, but I'm trying to extract an `OPTResource`, if it's present. I'm not sure how to do this without a type assertion.

Or did you mean without the type switch? Like this?
```go
var optRes *dnsmessage.OPTResource = nil
for _, additional := range msg.Additionals {
  p, ok := additional.Body.(OPTResource)
  if ok {
    optRes = p
  }
}
```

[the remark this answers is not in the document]
bemasc: Go's type switch appears to contain an invisible type assertion for each case, so you could do
```go
body := additional.Body
switch body.(type) {
case *dnsmessage.OPTResource:
    optRes = body
    break
}
```

However, it does seem like
```go
if optRes, ok := additional.Body.(*dnsmessage.OPTResource); ok {
    break
}
```
would be an even simpler solution.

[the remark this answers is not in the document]
dmcardle: Nevermind, I figured out what you mean.

```
switch body := additional.Body.(type) { ... }
```

tunnel/intra/doh/padding.go
dmcardle: You're right, it actually doesn't look too hard to use the `Builder` to control whether compression is enabled.

> I think the real goal is to test that this does the right thing when the input is not compressed (or is compressed differently from how dnsmessage would do it).

Absolutely.  To be clear, the "sanity check" tests on the behavior of the dnsmessage library are not the *only* tests.

I manually constructed the bytes of uncompressed and compressed queries, and wrote some tests using those.  I also have tests for queries with/without OPT and with/without padding.

Let me know if there's some situation I'm missing, and/or if you think I should explicitly use `Builder`.

tunnel/intra/doh/padding.go
```diff
@@ -0,0 +1,114 @@
...
+	} else {
+		// Create an empty OPTResource (contains no padding option) and
+		// push it into |msg.Additionals|.
+		optRes = &dnsmessage.OPTResource{
+			Options: []dnsmessage.Option{},
+		}
+		msg.Additionals = append(msg.Additionals, dnsmessage.Resource{
+			Header: dnsmessage.ResourceHeader{
+				Name:  dnsmessage.MustNewName("."),
+				Class: dnsmessage.ClassINET,
```
bemasc: Unfortunately, OPT has some [crazy ideas](https://tools.ietf.org/html/rfc6891#section-6.1.2) about what's supposed to go in these fields.  I think the "right way" is to call [SetEDNS0](https://godoc.org/golang.org/x/net/dns/dnsmessage#ResourceHeader.SetEDNS0).  To start, I would try `udpPayloadLen = 65535`, `extRCode = 0`, and `dnssecOK = false`.

dmcardle: Ah, thanks for that. Done!

I also improved the integration test so it sends a padded query to the server as well. FYI, it turns out that HonestDNS's responses include padding regardless of the request's padding.

dmcardle:
Friendly ping :)

tunnel/intra/doh/doh_test.go
```diff
@@ -559,3 +684,134 @@ func TestAcceptOversize(t *testing.T) {
...
+	}
+	if len(padded)%PaddingBlockSize != 0 {
+		t.Errorf("AddEdnsPadding failed to correctly pad compressed query")
+	}
+}
+
+// Try to pad a query that already contains an OPT record, but no padding option.
+func TestAddEdnsPaddingCompressedOptQuery(t *testing.T) {
+	optQuery := simpleQuery
+	optQuery.Additionals = append(optQuery.Additionals,
```
bemasc: I think this is permitted to mutate `simpleQuery.Additionals`.  It might not actually be mutated due to the way the slice is allocated, but I would prefer an explicit `copy` just to be safe.

tunnel/intra/doh/padding.go
```diff
@@ -0,0 +1,114 @@
...
+		// not contain a padding option.
+	} else {
+		// Create an empty OPTResource (contains no padding option) and
+		// push it into |msg.Additionals|.
+		optRes = &dnsmessage.OPTResource{
+			Options: []dnsmessage.Option{},
+		}
+
+		optHeader := dnsmessage.ResourceHeader{}
+		optHeader.SetEDNS0(65535, dnsmessage.RCodeSuccess, false)
```
bemasc: I would appreciate a comment indicating what these arguments are, so the reader doesn't have to go look at the docs for SetEDNS0.

tunnel/intra/doh/doh_test.go
dmcardle: OK, now I'm explicitly copying `simpleQuery.Additionals` before appending to it.

tunnel/intra/doh/padding.go
dmcardle: I just pasted the function signature in as a comment, as I think that is easier to read than a prose description.

dmcardle:
Done, PTAL :)

dmcardle:
There's a conflict to resolve because `NewTransport()` appears to have a new parameter. I'd resolve it, but it looks like `TestNewTransport` is currently failing on master.

bemasc:
Thanks for the bug report.  The test failure should be fixed by #42.

dmcardle:
Just rebased on master to incorporate #42.

dmcardle:
(Tests are passing now!)

PR #43, opened 2020-01-10
Improve protection of DNS queries

bemasc:
Go on Android cannot determine the system's preferred DNS servers.
This is normally fine, because Resolver.PreferGo defaults to false,
but we need to set PreferGo to true when protection is enabled,
in order to protect the DNS query's UDP socket.

This change requires a Protector to indicate the list of DNS
servers, so that the Resolver can create a protected socket to
one of those servers.

### DIFF
--- a/tunnel/intra/doh/ipmap/ipmap.go
+++ b/tunnel/intra/doh/ipmap/ipmap.go
@@ -104,7 +104,7 @@ func (s *IPSet) Add(hostname string) {
 	// Don't hold the ipMap lock during blocking I/O.
 	resolved, err := s.r.LookupIPAddr(context.TODO(), hostname)
 	if err != nil {
-		log.Infof("Error resolving %s: %v", hostname, err)
+		log.Warnf("Failed to resolve %s: %v", hostname, err)
 	}
 	s.Lock()
 	for _, addr := range resolved {
--- a/tunnel/intra/doh/ipmap/ipmap.go
+++ b/tunnel/intra/doh/ipmap/ipmap.go
@@ -19,6 +19,8 @@ import (
 	"math/rand"
 	"net"
 	"sync"
+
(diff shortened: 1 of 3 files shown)


tunnel/intra/doh/ipmap/ipmap.go
```diff
@@ -100,7 +102,10 @@ func (s *IPSet) add(ip net.IP) {
 // The hostname can be a domain name or an IP address.
 func (s *IPSet) Add(hostname string) {
 	// Don't hold the ipMap lock during blocking I/O.
-	resolved, _ := s.r.LookupIPAddr(context.TODO(), hostname)
+	resolved, err := s.r.LookupIPAddr(context.TODO(), hostname)
+	if err != nil {
+		log.Infof("Error resolving %s: %v", hostname, err)
```
alalamav: nit: Errorf?

tunnel/intra/protect/protect.go
```diff
@@ -36,6 +47,34 @@ func makeControl(p Protector) (func(string, string, syscall.RawConn) error) {
 	}
 }
 
+// Returns the first IP address that is of the specified length (i.e. family).
+func scan(ips []string, l int) string {
```
alalamav: Using the address length may be error prone. Can we use `net.ParseIP(ip).To4() != nil` to determine the address family?

tunnel/intra/doh/ipmap/ipmap.go
bemasc: I changed this to `Warnf`.  (I try to reserve "Error" for bugs, whereas this error will occur whenever you are experiencing DNS poisoning of a resolver name.)

tunnel/intra/protect/protect.go
bemasc: OK, done.

PR #46, opened 2020-02-12
Add a tool for testing split-handshake behavior

bemasc:


tunnel/intra/split/example/main.go
```diff
@@ -0,0 +1,72 @@
...
+	"crypto/tls"
+	"log"
+	"net"
+	"os"
+
+	"github.com/Jigsaw-Code/outline-go-tun2socks/tunnel/intra/split"
+)
+
+func main() {
+	if len(os.Args) < 2 {
```
alalamav: Consider using the `flags` module for better type safety and documentation.

tunnel/intra/split/example/main.go
```diff
@@ -0,0 +1,72 @@
...
+		log.Printf("Usage: %s destination [SNI]", os.Args[0])
+		log.Printf("This tool attempts TLS connection to the destination (port 443), with and without splitting.")
+		log.Printf("If the SNI is specified, it overrides the destination, which can be an IP address.")
+		return
+	}
+	destination := os.Args[1]
+	addr, err := net.ResolveTCPAddr("tcp", net.JoinHostPort(destination, "443"))
+	if err != nil {
+		log.Println("Couldn't resolve destination")
```
alalamav: Use `log.Fatal` to avoid the return statement and exit with non-zero status code.

tunnel/intra/split/example/main.go
bemasc: Done

tunnel/intra/split/example/main.go
bemasc: Done

PR #47, opened 2020-03-03
Don't panic if protect() returns false

bemasc:
This happens frequently on Kitkat (Android 4.4).
A panic might be appropriate if protect _always_ fails,
because the whole system will be nonfunctional, but
for sporadic or edge-case failures it's ok to log
the error and continue.


bemasc:
> Do we want to keep track of the frequency of these failures so we can panic if appropriate?

I added a TODO to track these errors.  However, I don't think we should call `panic()` even if we see a lot of these errors.  I think of `panic()` as meaning "we believe this code is not reachable".

PR #50, opened 2020-03-31
Add domain error reporting with Choir

bemasc:


tunnel/intra/sni_reporter.go
```diff
@@ -0,0 +1,116 @@
...
+// Send implements choir.ReportSender.
+func (r *tcpSNIReporter) Send(report choir.Report) error {
+	r.mu.RLock()
+	suffix := r.suffix
+	dns := r.dns
+	r.mu.RUnlock()
+	q, err := choir.FormatQuery(report, suffix)
+	if err != nil {
+		log.Warnf("Failed to construct query for Choir: %v", err)
+		return nil
```
alalamav, line 64: Should you return `err`? Same below.

tunnel/intra/sni_reporter.go
```diff
@@ -0,0 +1,116 @@
...
+	if summary.DownloadBytes > 0 {
+		result = "success"
+	}
+	response := "closed"
+	if summary.Retry.Timeout {
+		response = "timeout"
+	}
+	resultValue, err := choir.NewValue(result)
+	if err != nil {
+		panic(err)
```
alalamav, line 107: Would it be useful to return the error instead? It seems like a error in reporting should not crash the entire app.

tunnel/intra/sni_reporter.go
```diff
@@ -0,0 +1,116 @@
...
+		log.Infof("Failed to deliver query for Choir: %v", err)
+	}
+	return nil
+}
+
+// Configure initializes or reinitializes the reporter.
+// `file` is the Choir salt file (persistent and initially empty).
+// `suffix` is the domain to which reports will be sent.
+// `country` is the two-letter ISO country code of the user's location.
+func (r *tcpSNIReporter) Configure(file io.ReadWriter, suffix, country string) (err error) {
```
alalamav, line 76: Will the configuration change often after the object is created? If not, this could be renamed `NewSNIReporter` and the object can become immuatable.

alalamav:
nit: it would be useful add the Choir link to the `README` or `tunnel/intra/sni_reporter.go`.

tunnel/intra/sni_reporter.go
bemasc: The only action Choir would take on an error is to print it using the standard Go logging framework: https://github.com/Jigsaw-Code/choir/blob/master/client.go#L105.  Since we have our own logging framework here, I figured it would be better to log here and not there.

alalamav: Makes sense 👍

tunnel/intra/sni_reporter.go
alalamav: The [golang readability](https://g3doc.corp.google.com/go/g3doc/readability/review-comments.md?cl=head#dont-panic) suggests using `log.Fatal`. The error log could clarify that this is an "impossible" failure or you can add a comment.

bemasc: OK, done

PR #52, opened 2020-06-16
Enable go modules

alalamav:
- Enables building all targets with go modules.
  - Removes symlinking to modularize source packages.
- Updates to go v1.14.
- Updates dependencies
- Migrates to `techknowlogick/xgo`, which supports cross compiling (CGO) go modules and is now more actively maintained than the parent repository.


bemasc:
This looks great, but I just tried it locally and got a flood of these errors after running `./build_android.sh intra`:
```
rm: cannot remove '/tmp/gomobile-work-174985860/pkg/mod/github.com/beorn7/perks@v1.0.1/topk/topk_test.go': Permission denied
rm: cannot remove '/tmp/gomobile-work-174985860/pkg/mod/github.com/beorn7/perks@v1.0.1/topk/topk.go': Permission denied
rm: cannot remove '/tmp/gomobile-work-174985860/pkg/mod/github.com/beorn7/perks@v1.0.1/LICENSE': Permission denied
...
```

alalamav:
Ahh, I think we're getting these errors because now packages are copied to the work directory straight from the modules directory, which may have root permissions?

What do you think about not removing the work directory? It should be fine since it will be located on `/tmp` almost always.

bemasc:
> Ahh, I think we're getting these errors because now packages are copied to the work directory straight from the modules directory, which may have root permissions?

It looks like the issue is that the files and directories don't have the `+w` mode, and `rm` can't delete files from a read-only directory.  `chmod -R +w $GO_WORK_DIR` seems to do the trick...

alalamav:
I implemented your suggestion to make the Go working directory writable.

PR #56, opened 2020-07-16
Concatenate the Shadowsocks header and payload

bemasc:


bemasc:
See https://github.com/Jigsaw-Code/outline-ss-server/pull/73

PR #58, opened 2020-08-04
Run tests in presubmit

bemasc:


bemasc:
Yes, `go build` works without gomobile.  CGO is built-in to Go, I think.  It builds everything, and then runs all the tests: https://github.com/Jigsaw-Code/outline-go-tun2socks/pull/58/checks?check_run_id=946312536

I think this all works because we don't have any reverse bindings, so the Go code doesn't depend on any platform code.

PR #62, opened 2020-09-16
Close the TUN writer when shutting down the tunnel

bemasc:
Otherwise, the TUN writer is only closed by the garbage collector, which
can cause a long delay.

See https://github.com/Jigsaw-Code/Intra/issues/285

### DIFF
--- a/tunnel/tunnel.go
+++ b/tunnel/tunnel.go
@@ -47,6 +47,7 @@ func (t *tunnel) Disconnect() {
 	}
 	t.isConnected = false
 	t.lwipStack.Close()
+	t.tunWriter.Close()
 }
 
 func (t *tunnel) Write(data []byte) (int, error) {


bemasc:
@alalamav This change impacts Outline, but I haven't tested this with Outline on any platform, and I'm not familiar with the TUN writer ownership in those cases, so let me know if you think this is safe and correct.

alalamav:
In Outline, Go [does](https://github.com/Jigsaw-Code/outline-go-tun2socks/blob/master/outline/android/tun2socks.go#L38) [not](https://github.com/Jigsaw-Code/outline-client/blob/go-tun2socks/cordova-plugin-outline/android/java/org/outline/vpn/VpnTunnel.java#L109) take ownership of the TUN file descriptor.  Is there a reason Go needs to take ownership in Intra? I realize you [changed](https://github.com/Jigsaw-Code/Intra/pull/283) this behavior in Intra; we should probably keep it consistent either way.

alalamav:
One argument in favor of keeping the TUN ownership in Java is to support network configuration changes without tearing down the VPN.

bemasc:
Interesting.  I think that comment is currently incorrect: Go will close the VPN FD when the corresponding File object is garbage-collected.  In other words, `os.NewFile()` takes ownership of the file descriptor.  There is [no way to release](https://github.com/golang/go/issues/24215) ownership when using `os.File`, so if we want that behavior, we'd have to switch to [unix.Write](https://godoc.org/golang.org/x/sys/unix#Write) (and [windows.Write](https://godoc.org/golang.org/x/sys/windows#Write)?).  (Either that, or somehow let the `os.File` outlive the `Tunnel`.)

Do we have a scenario where we want to reconfigure the tunnel without recreating the TUN device?

alalamav:
Oh I see, thanks for explaining.

> Do we have a scenario where we want to reconfigure the tunnel without recreating the TUN device?

In Outline, we support connecting to a different server without tearing down the VPN to avoid traffic leaks. This problem only applies to Android, since other platforms handle the TUN fd at different levels. Android supports [seamless VPN handover](https://developer.android.com/reference/android/net/VpnService.Builder#establish()), so we could re-establish the VPN before disconnecting the tunnel. 

Ideally, I would prefer to keep the tunnel behavior separate from the VPN, so maybe we could implement a `tunWriter` that uses `unix.Write` instead of `os.NewFile`.

We may also retain the `file` object, but that may complicate the Go API.

bemasc:
OK, I've implemented that in #63.

PR #63, opened 2020-09-17
Don't use os.NewFile to create the TUN writer

bemasc:
os.NewFile takes ownership of the file descriptor, and adds a finalizer
that closes the file descriptor when the file is garbage-collected.
This violates Outline's interface contract (caller retains ownership of
the TUN device), and could cause a problem if the file descriptor is
still being used

See https://github.com/Jigsaw-Code/Intra/issues/285

### DIFF
--- a/tunnel/tun.go
+++ b/tunnel/tun.go
@@ -1,40 +1,46 @@
 package tunnel
 
 import (
-	"errors"
-	"os"
+	"io"
 
 	"github.com/eycorsican/go-tun2socks/common/log"
 	_ "github.com/eycorsican/go-tun2socks/common/log/simple" // Import simple log for the side effect of making logs printable.
+	"golang.org/x/sys/unix"
 )
 
 const vpnMtu = 1500
 
-// MakeTunFile returns an os.File object from a TUN file descriptor `fd`.
-func MakeTunFile(fd int) (*os.File, error) {
-	if fd < 0 {
-		return nil, errors.New("Must provide a valid TUN file descriptor")
-	}
-	file := os.NewFile(uintptr(fd), "")
-	if file == nil {
-		return nil, errors.New("Failed to open TUN file descriptor")
-	}
-	return file, nil
+// TUNFile is an io.ReadWriter wrapping a TUN file descriptor (UNIX platforms only).
+// This is a substitute for os.NewFile that avoids taking ownership of the file.
+type TUNFile int
+
+func (f TUNFile) Read(buf []byte) (int, error) {
+	return unix.Read(int(f), buf)
+}
+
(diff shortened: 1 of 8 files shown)


tunnel/tun.go
```diff
@@ -1,40 +1,46 @@
...
-		return nil, errors.New("Must provide a valid TUN file descriptor")
-	}
-	file := os.NewFile(uintptr(fd), "")
-	if file == nil {
-		return nil, errors.New("Failed to open TUN file descriptor")
-	}
-	return file, nil
+// TUNFile is an io.ReadWriter wrapping a TUN file descriptor (UNIX platforms only).
+// This is a substitute for os.NewFile that avoids taking ownership of the file.
+type TUNFile int
```
alalamav, line 15: Is it worth defining the wrapper as a struct with `fd` as a member?

tunnel/tun.go
```diff
@@ -1,40 +1,46 @@
...
+
+func (f TUNFile) Read(buf []byte) (int, error) {
```
alalamav, line 17: Would it be possible to check whether the `fd` is valid so we can fail fast instead of on first read/write? For instance calling [fcntl](https://man7.org/linux/man-pages/man2/fcntl.2.html) on init?

alalamav (commented):
Thanks for the improvement and refactoring the electron binary! I have some questions and want to do a quick test on Outline before approving.

bemasc:
After further thought, I don't think this will work on Android pre-21 (L), because the VPN file descriptor will be non-blocking.  That's probably not a problem for writing to the TUN device, but for reading from the device it would cause our `ProcessInputPackets` to spin until a packet arrives.  It looks like the right answer is to use `syscall.RawConn`, which is similar to this arrangement but not the same.

I will try to update this PR to use `RawConn`.  In the meantime, I've marked it as "draft".

alalamav:
Good catch. Is it possible to use a syscall to set the fd to blocking mode (API <21) or is it not supported by the OS?
Outline Android minimum supported version is 22, so we could split the tunWriter implementation by platfrom, perhaps even the ownership contract.

tunnel/tun.go
bemasc: (obsolete)

tunnel/tun.go
bemasc: (obsolete)

bemasc:
Update: I've reworked this to use `unix.Dup()` to add a separate reference to the TUN device (matching a suggestion by @JonathanDCohen).  This allows us to avoid reimplementing the Go stdlib's file-handling code, and also provides a fairly natural reference-counting behavior for file management.

alalamav (approved):
LGTM!

PR #70, opened 2020-12-08
Separate Intra and Outline packages

bemasc:
This slightly reduces the AAR size for the Intra and Outline Android
builds (-1% for Intra, -0.2% for Outline).  More importantly, it
provides a clearer separation of functionality into package trees.

This is primarily a code move, but does make some small changes to
the `outlinetunnel` and `intratunnel` structs to avoid relying on private
types from the `tunnel` package.


alalamav (approved):
LGTM - should we rename `[Intra|Outline]Tunnel` to `Tunnel` since they are now in their respective package?

bemasc:
Done!

PR #71, opened 2020-12-14
Remove example binary and add it to gitignore

bemasc:


.gitignore
```diff
@@ -1,12 +1,13 @@
...
 
 # General
 .DS_Store
 .AppleDouble
 .LSOverride
 
 # Icon must end with two \r
-Icon

+Icon
```
alalamav, line 10: Is the comment accurate?

alalamav (approved):

.gitignore
bemasc: Hm, I don't know what this is about.  It looks like my editor did this by itself.  Only vim seems to preserve this weird line ending.  (Fixed.)

Issue #72, opened 2020-12-28
udp.go:udpHandler.fetchUDPInput timeout too high


ignoramous:
At least on two occasions we've noticed _very_ _many_ `udp.go:fetchUDPInput` go routines block at [`net.UDPConn.ReadFrom`](https://github.com/Jigsaw-Code/outline-go-tun2socks/blob/7332825429/intra/udp.go#L100) for more than 60 minutes. Not unexpected since the [current timeout is set to `2h4m`](https://github.com/Jigsaw-Code/outline-go-tun2socks/blob/304acda35e1/intra/tunnel.go#L93) but unsure if such a long idle timeout is req for UDP connections, too. Isn't it the case that most routers close UDP NAT after idling for `30s`? If so, `2h4m` idle timeout might be way too high. [RFC-4787](https://tools.ietf.org/html/rfc4787) I stumbled upon, if I read it correctly, recommends at least 2mins or 5mins timeouts for UDP NAT.

One such io-wait:

```
12-27 20:46:09.819  9777     0 E Go      : goroutine 8785 [IO wait, 72 minutes]:
12-27 20:46:09.819  9777     0 E Go      : internal/poll.runtime_pollWait(0x7f7e0b7158, 0x72, 0x0)
12-27 20:46:09.819  9777     0 E Go      : >/home/murtaza/.gvm/gos/go1.15/src/runtime/netpoll.go:220 +0x3c
12-27 20:46:09.820  9777     0 E Go      : internal/poll.(*pollDesc).wait(0x4000571d18, 0x72, 0x800, 0x800, 0x0)
12-27 20:46:09.820  9777     0 E Go      : >/home/murtaza/.gvm/gos/go1.15/src/internal/poll/fd_poll_runtime.go:87 +0x30
12-27 20:46:09.820  9777     0 E Go      : internal/poll.(*pollDesc).waitRead(...)
12-27 20:46:09.820  9777     0 E Go      : >/home/murtaza/.gvm/gos/go1.15/src/internal/poll/fd_poll_runtime.go:92
12-27 20:46:09.820  9777     0 E Go      : internal/poll.(*FD).ReadFrom(0x4000571d00, 0x40002f0000, 0x800, 0x800, 0x0, 0x0, 0x0, 0x0, 0x0)
12-27 20:46:09.820  9777     0 E Go      : >/home/murtaza/.gvm/gos/go1.15/src/internal/poll/fd_unix.go:215 +0x154
[...]

PR #73, opened 2020-12-28
Set Intra UDP NAT timeout to 5 minutes

ignoramous:
A significant amount of UDP connections on Android are DNS requests.
An hour long timeout for them isn't really helpful especially considering DNS
TTLs themselves are considerably shorter. Besides, RFC 4787 recommends
at least a timeout of 5mins, which is still high for DNS, but may work nicely for
other UDP connections, at least according to the RFC.

#72


ignoramous:
@googlebot I signed it!

ignoramous:
cc: @bemasc

bemasc (approved):

bemasc:
You're right; we were applying the TCP timeout to UDP.  Note that DNS is generally not affected: the DNS handler closes the mapping as soon as the response is relayed.

PR #76, opened 2021-01-26
Shadowsocks online config

alalamav:
- Implements SIP008 online config according to the [published spec](https://github.com/shadowsocks/shadowsocks-org/wiki/SIP008-Online-Configuration-Delivery).
  - Outline-specific policies, such as rejecting unsupported configurations, and error handling will be implemented in `outline/shadowsocks`.
- Supports HTTP redirects, self-signed certificates with validation through the certificate SHA256 fingerprint, and specifying the HTTP method to use.
- Implements end-to-end tests.

### DIFF
--- a/shadowsocks/config.go
+++ b/shadowsocks/config.go
@@ -16,13 +16,11 @@ package shadowsocks
 
 import (
 	"bytes"
-	"context"
 	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
 	"errors"
-	"net"
 	"net/http"
 	"time"
 )
@@ -124,8 +122,6 @@ func FetchConfig(req FetchConfigRequest) (*FetchConfigResponse, error) {
 	res.Config = config
 	return &res, err
 }
-
-type tlsDialer func(ctx context.Context, network, addr string) (net.Conn, error)
 
 type certVerifier func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error
 
--- a/shadowsocks/config.go
+++ b/shadowsocks/config.go
@@ -106,7 +106,7 @@ func FetchOnlineConfig(req OnlineConfigRequest) (*OnlineConfigResponse, error) {
 	}
 	defer func() {
 		// Read the response body to EOF and close it, ignoring errors.
-		_, _ = ioutil.ReadAll(httpres.Body)
+		ioutil.ReadAll(httpres.Body)
 		httpres.Body.Close()
 	}()
 
--- a/shadowsocks/config.go
+++ b/shadowsocks/config.go
@@ -1,0 +1,149 @@
+// Copyright 2021 The Outline Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package shadowsocks
+
+import (
+	"context"
+	"crypto/sha256"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"net"
+	"net/http"
+	"time"
+)
+
+// ProxyConfig represents a Shadowsocks proxy configuration.
+type ProxyConfig struct {
+	Host       string `json:"server"`
+	Port       int    `json:"server_port"`
+	Password   string `json:"password"`
+	Cipher     string `json:"method"`
+	Name       string `json:"remarks,omitempty"`
+	Plugin     string `json:"plugin,omitempty"`
+	PluginOpts string `json:"plugin_opts,omitempty"`
+}
+
+// FetchConfigRequest encapsulates a request to an online config server.
+type FetchConfigRequest struct {
+	// URL is the HTTPs endpoint of an online config server.
+	URL string
+	// Method is the HTTP method to use in the request.
+	Method string
+	// TrustedCertFingerprint is the base64-encoded sha256 hash of the online
+	// config server's TLS certificate.
+	TrustedCertFingerprint string
+}
+
+// FetchConfigResponse encapsulates a response and metadata from an online config server.
+type FetchConfigResponse struct {
+	// Proxies is a list of Shadowsocks proxy configurations
+	Proxies []ProxyConfig
+	// HTTPStatusCode is the HTTP status code of the response.
+	HTTPStatusCode int
+	// RedirectURL is the Location header of a HTTP redirect response.
+	RedirectURL string
+}
+
+// sip008Response represents a JSON response from an online config server.
+type sip008Response struct {
+	Proxies []ProxyConfig `json:"servers"`
+}
+
+// FetchConfig retrieves Shadowsocks proxy configurations per SIP008:
+// https://github.com/shadowsocks/shadowsocks-org/wiki/SIP008-Online-Configuration-Delivery
+//
+// Pins the trusted certificate when req.TrustedCertFingerprint is non-empty.
+// Sets the response's RedirectURL when the status code is a redirect.
+// Returns an error if req.URL is a non-HTTPS URL, if there is a connection
+// error to the server, or if parsing the configuration fails.
+func FetchConfig(req FetchConfigRequest) (*FetchConfigResponse, error) {
+	httpreq, err := http.NewRequest(req.Method, req.URL, nil)
+	if err != nil {
+		return nil, err
+	}
+	if httpreq.URL.Scheme != "https" {
+		return nil, errors.New("URL protocol must be HTTPs")
+	}
+
+	client := &http.Client{
+		// Do not follow redirects automatically, relay to the caller.
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			return http.ErrUseLastResponse
+		},
+		Timeout: 30 * time.Second,
+	}
+
+	if req.TrustedCertFingerprint != "" {
+		client.Transport = &http.Transport{
+			DialTLSContext: makePinnedCertTLSDialer(req.TrustedCertFingerprint),
+		}
+	}
+
+	httpres, err := client.Do(httpreq)
+	if err != nil {
+		return nil, err
+	}
+
+	var res FetchConfigResponse
+	res.HTTPStatusCode = httpres.StatusCode
+	if res.HTTPStatusCode >= 300 && res.HTTPStatusCode < 400 {
+		// Redirect
+		res.RedirectURL = httpres.Header.Get("Location")
+		return &res, nil
+	} else if res.HTTPStatusCode > 400 {
+		// HTTP error
+		return &res, nil
+	}
+
+	// 2xx status code
+	defer httpres.Body.Close()
+	var sip008res sip008Response
+	err = json.NewDecoder(httpres.Body).Decode(&sip008res)
+	res.Proxies = sip008res.Proxies
+	return &res, err
+}
+
+type tlsDialer func(ctx context.Context, network, addr string) (net.Conn, error)
+
+// Returns a dial TLS context that pins trustedCertFingerprint for certificate
+// validation. Trusts the connection if the server certificate fingerprint
+// matches the pinned certificate fingerprint, regardless of the system's
+// TLS certificate validation errors.
+func makePinnedCertTLSDialer(trustedCertFingerprint string) tlsDialer {
+	return func(ctx context.Context, network, addr string) (net.Conn, error) {
+		c, err := tls.Dial(network, addr, &tls.Config{InsecureSkipVerify: true})
+		if err != nil {
+			return c, err
+		}
+		connState := c.ConnectionState()
+		for _, cert := range connState.PeerCertificates {
+			fingerprint := computeCertificateFingerprint(cert)
+			if fingerprint == trustedCertFingerprint {
+				return c, nil
+			}
+		}
+		return nil, errors.New("Failed to validate TLS certificate")
+	}
+}
+
+// Computes the sha256 digest of the whole DER-encoded certificate and
+// returns it as a base64-encoded string.
+func computeCertificateFingerprint(cert *x509.Certificate) string {
+	digest := sha256.Sum256(cert.Raw)
+	return base64.StdEncoding.EncodeToString(digest[:])
+}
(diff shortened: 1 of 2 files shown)


shadowsocks/config.go
```diff
@@ -0,0 +1,149 @@
...
+	Proxies []ProxyConfig
+	// HTTPStatusCode is the HTTP status code of the response.
+	HTTPStatusCode int
+	// RedirectURL is the Location header of a HTTP redirect response.
+	RedirectURL string
+}
+
+// sip008Response represents a JSON response from an online config server.
+type sip008Response struct {
+	Proxies []ProxyConfig `json:"servers"`
```
bemasc, line 64: I think this also needs a "version" field.

shadowsocks/config.go
```diff
@@ -0,0 +1,149 @@
...
+	client := &http.Client{
+		// Do not follow redirects automatically, relay to the caller.
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			return http.ErrUseLastResponse
+		},
+		Timeout: 30 * time.Second,
+	}
+
+	if req.TrustedCertFingerprint != "" {
+		client.Transport = &http.Transport{
```
bemasc, line 92: I think the standard way to do this is by overriding `http.Transport.TLSClientConfig.VerifyPeerCertificate`.

shadowsocks/config.go
```diff
@@ -0,0 +1,149 @@
...
+		// Redirect
+		res.RedirectURL = httpres.Header.Get("Location")
+		return &res, nil
+	} else if res.HTTPStatusCode > 400 {
+		// HTTP error
+		return &res, nil
+	}
+
+	// 2xx status code
+	defer httpres.Body.Close()
```
bemasc, line 114: Any reason not to do this earlier?

shadowsocks/config.go
```diff
@@ -0,0 +1,149 @@
...
+	PluginOpts string `json:"plugin_opts,omitempty"`
+}
+
+// FetchConfigRequest encapsulates a request to an online config server.
+type FetchConfigRequest struct {
+	// URL is the HTTPs endpoint of an online config server.
+	URL string
+	// Method is the HTTP method to use in the request.
+	Method string
+	// TrustedCertFingerprint is the base64-encoded sha256 hash of the online
```
bemasc, line 47: Currently, it looks like you're storing the fingeprint as base64, computing the new fingerprint in base64, and comparing them as strings.  This seems surprising to me.  Can we make this a `[]byte`, and do the comparison on the binary hashes?

shadowsocks/config_test.go
```diff
@@ -0,0 +1,260 @@
...
+	exampleCertFingerprint = "IA3K+nZ8hFDs5kSHnAYqDN9SJA/gW7frKEYRw67z7C4="
+)
+
+var proxies = []ProxyConfig{
+	{"ssconf.test", 123, "passw0rd", "chacha20-ietf-poly1305", "ssconf-test-1", "", ""},
+	{"ssconf-ii.test", 456, "dr0wssap", "chacha20-ietf-poly1305", "ssconf-test-2", "", ""},
+}
+
+func TestFetchConfig(t *testing.T) {
+	serverAddr := "127.0.0.1:9999"
```
bemasc, line 84: In general, it's best to avoid hardcoding port numbers in tests.  Using port "0", which causes the OS to choose an open port for you, works better.  Otherwise you run this risk of test failures if the port is not available (e.g. if multiple instances of the test are running concurrently).

shadowsocks/config_test.go
```diff
@@ -0,0 +1,260 @@
...
+	now := time.Now()
+	template := &x509.Certificate{
+		SerialNumber: big.NewInt(now.Unix()),
+		Subject: pkix.Name{
+			Organization: []string{"online config"},
+		},
+		IPAddresses:           []net.IP{net.IPv4(127, 0, 0, 1)}, // Valid for localhost
+		NotBefore:             now,
+		NotAfter:              now.AddDate(0, 0, 1), // Valid for one day
+		SubjectKeyId:          []byte{55, 43, 04, 45, 87, 65},
```
bemasc, line 213: What is this?

shadowsocks/config_test.go
alalamav: Done in afd897e.

shadowsocks/config.go
alalamav: Done in eebc114. Good suggestion.

shadowsocks/config.go
bemasc: I think you're required to close it anyway.  In the early return cases, the body is not closed here, which is [apparently bad](https://golang.org/pkg/net/http/#Client.Do):
> If the Body is not both read to EOF and closed, the Client's underlying RoundTripper (typically Transport) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.

shadowsocks/config.go
```diff
@@ -0,0 +1,148 @@
...
+	"net/http"
+	"time"
+)
+
+// Config represents a Shadowsocks proxy configuration.
+type Config struct {
+	ID         string
+	Host       string `json:"server"`
+	Port       int    `json:"server_port"`
+	Password   string `json:"password"`
```
bemasc, line 33: Is this JSON annotation redundant?

shadowsocks/config.go
```diff
@@ -0,0 +1,148 @@
...
+	// HTTPStatusCode is the HTTP status code of the response.
+	HTTPStatusCode int
+	// RedirectURL is the Location header of a HTTP redirect response.
+	RedirectURL string
+}
+
+// OnlineConfig represents a SIP008 response from an online config server.
+type OnlineConfig struct {
+	Proxies []Config `json:"servers"`
+	Version int
```
bemasc, line 64: Something should probably check this version code and abort if it's not `1`.

shadowsocks/config.go
```diff
@@ -0,0 +1,148 @@
...
+	Method string
+	// TrustedCertFingerprint is the sha256 hash of the online config server's
+	// TLS certificate.
+	TrustedCertFingerprint []byte
+}
+
+// FetchConfigResponse encapsulates a response from an online config server.
+type FetchConfigResponse struct {
+	// Config is the parsed server response.
+	Config OnlineConfig
```
bemasc, line 54: I feel like this is confusing `FetchConfigResponse.Config` is not a `Config`), especially because there's a singular/plural mismatch (it contains a `[]Config`).

shadowsocks/config.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+type certVerifier func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error
+
+// Verifies whether the pinned certificate SHA256 fingerprint,
+// trustedCertFingerprint, matches a fingerprint in the certificate chain,
+// regardless of the system's TLS certificate validation errors.
+func makePinnedCertVerifier(trustedCertFingerprint []byte) certVerifier {
+	return func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+		for _, cert := range rawCerts {
```
bemasc, line 133: If `InsecureSkipVerify` is true, does `crypto/tls` still require that the certificate chain is internally consistent (i.e. each certificate signs the preceding one)?  If not, an adversary could simply copy-paste the true certificate into their cert chain to pass this check.

I see a few options:
1. Just check the fingerprint of `rawCerts[0]`.  I don't see why we would need to support chaining.
2. Scan `rawCerts` for a cert matching the fingerprint (`rootCert`), and do `leafCert.Verify` using `rawCerts` as intemediates and `rootCert` as the root CA.

shadowsocks/config.go
```diff
@@ -0,0 +1,148 @@
...
+				return nil
+			}
+		}
+		return x509.CertificateInvalidError{
+			nil, x509.NotAuthorizedToSign, "Failed to validate pinned TLS certificate"}
+	}
+}
+
+// Computes the sha256 digest of the whole DER-encoded certificate
+func computeCertificateFingerprint(derCert []byte) []byte {
```
bemasc, line 145: I'm not sure this function is still worthwhile.

shadowsocks/config_test.go
```diff
@@ -0,0 +1,332 @@
...
+		return tls.Certificate{}, err
+	}
+
+	var cert tls.Certificate
+	cert.Certificate = append(cert.Certificate, derCert)
+	cert.PrivateKey = key
+	return cert, nil
+}
+
+func TestComputeCertificateFingerprint(t *testing.T) {
```
bemasc, line 317: This also seems a bit unnecessary now.

shadowsocks/config.go
```diff
@@ -0,0 +1,148 @@
...
+	Port       int    `json:"server_port"`
+	Password   string `json:"password"`
+	Cipher     string `json:"method"`
+	Name       string `json:"remarks,omitempty"`
+	Plugin     string `json:"plugin,omitempty"`
+	PluginOpts string `json:"plugin_opts,omitempty"`
+}
+
+// FetchConfigRequest encapsulates a request to an online config server.
+type FetchConfigRequest struct {
```
bemasc, line 41: Maybe just `ConfigRequest`/`ConfigResponse`?

shadowsocks/config.go
alalamav: Yes, removed.

shadowsocks/config.go
alalamav: I was trying to be lenient in the version we accept. My thinking was that that if there is a breaking change, decoding would fail anyway. If we want to surface this particular error to the user, we can check the version in the method we expose to the client - wdyt?

shadowsocks/config.go
alalamav: Done - please see my comment below.

shadowsocks/config.go
alalamav: I further qualified the model: 
- `Config` -> `ProxyConfig`
- `FetchConfig[Request|Response]` -> `OnlineConfig[Request|Response]`
- `FetchConfig` -> `FetchOnlineConfig`
- `FetchConfigResponse.Config` -> `OnlineConfigResponse.OnlineConfig`

Please let me know if this makes sense to you.

alalamav: Another option to address the singular/plural mismatch would be to embed `OnlineConfig` in `OnlineConfigResponse`. That way, `OnlineConfigResponse.Proxies` would be consistent.

shadowsocks/config.go
bemasc: Hm, I guess it's not clear what this version code is for.  A backwards-compatible version code isn't very useful, but the spec doesn't explain what to do with it.  Maybe just log a warning if it's not `1`?

shadowsocks/config.go
```diff
@@ -0,0 +1,145 @@
...
+
+type certVerifier func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error
+
+// Verifies whether the pinned certificate SHA256 fingerprint,
+// trustedCertFingerprint, matches a fingerprint in the certificate chain,
+// regardless of the system's TLS certificate validation errors.
+func makePinnedCertVerifier(trustedCertFingerprint []byte) certVerifier {
+	return func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+		// Compute the sha256 digest of the whole DER-encoded certificate
+		fingerprint := sha256.Sum256(rawCerts[0])
```
bemasc, line 138: Is it possible to crash this code by sending a ServerHello with no certificates?

shadowsocks/config.go
bemasc: The latest version looks good to me.

shadowsocks/config.go
bemasc: Works for me.

bemasc (approved):

shadowsocks/config.go
alalamav: I would think the connection times out, but I added a check to be safe.

shadowsocks/config.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// FetchOnlineConfig retrieves Shadowsocks proxy configurations per SIP008:
+// https://github.com/shadowsocks/shadowsocks-org/wiki/SIP008-Online-Configuration-Delivery
+//
+// Pins the trusted certificate when req.TrustedCertFingerprint is non-empty.
+// Sets the response's RedirectURL when the status code is a redirect.
+// Returns an error if req.URL is a non-HTTPS URL, if there is a connection
+// error to the server, or if parsing the configuration fails.
+func FetchOnlineConfig(req OnlineConfigRequest) (*OnlineConfigResponse, error) {
```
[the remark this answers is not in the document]
fortuna: I would also want us to separate how we follow redirects and fetch the page from how we process the contents. Those can be defined separately and have their own nuances. Mixing them together makes it harder to understand and change.

We need some layers:
1) Fetch Access Code config: given an access code, returns a simple Outline proxy config (an Outline concept, not a SS concept) and the new access code, when applicable. This will run different logic whether it's a ss:// url or a ssconf/https.
2) Parse ss link: will simply parse the ss link and return the Outline config
3) Fetch online config
    1. given a ssconf/https url, extract the fingerprint, issue the HTTPS request (verifying the fingerprint).
    2. Parse the SS config, convert to the Outline config model and return if no redirect
    3. If redirected, save permanent redirect and call FetchAccessCode on the new url. Return that result, adding the new access code if absent from the response
4) Fetch HTTPS with fingerprint, which takes a http.Request + FP, does one simple roundtrip and returns the http response. Used by Fetch online config

I expressed as a recursion, but you can express it as a loop, so you avoid stack growth.

You also need to handle possible redirect cycles to prevent malicious access codes from DDoSing others.

fortuna (changes requested):

shadowsocks/config.go
```diff
@@ -0,0 +1,154 @@
...
+		res.RedirectURL = httpres.Header.Get("Location")
+		return &res, nil
+	} else if res.HTTPStatusCode >= 400 {
+		// HTTP error
+		return &res, nil
+	}
+
+	var config OnlineConfig
+	err = json.NewDecoder(httpres.Body).Decode(&config)
+	if config.Version != 1 {
```
fortuna, line 128: I'd like to see a function that all it does is to map the ss config to the Outline config model, so it's clear how we interpret it and we have something to point others to.

fortuna (changes requested):

shadowsocks/config.go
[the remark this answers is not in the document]
alalamav: I can refactor Transport creation (cert pinning) into a separate file.

Regarding layering: my intention is for this file to implement SIP008 per the spec (read: ss concepts). I am working on implementing Outline-specific concepts in the `outline/shadowsocks/` package, where we will expose library functions.

As for redirects: why do we need to support ss:// URL redirection? The online config spec requires JSON for the final delivery. I like the idea of automatically following redirects (parsing ssconf:// URLs if needed) and returning the final URL for permanent redirects.

Also, are you suggesting to implement proxy configuration selection policy in Go? I think it may be useful to return the proxy config list to the web app so we can implement retry logic there.

shadowsocks/config.go
alalamav: I'd like to implement this and config parameter validation in the `outline/shadowsocks` package.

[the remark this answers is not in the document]
alalamav: The implementation does not require that the version is present; it does, however, require that it is a number.

PR #79, opened 2021-03-05
Add an Outline test tool for linux

bemasc:
This allows rapid testing and profiling of outline-go-tun2socks
changes on linux, without needing to use the graphical client.

### DIFF
--- a/outline/electron/connect_linux.sh
+++ b/outline/electron/connect_linux.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# This script allows debugging outline-go-tun2socks directly on linux.
+# Instructions:
+# 1. Install the Outline client for Linux, connect to a server, and disconnect.
+#    This installs the outline controller service.
+# 2. $ git update-index --assume-unchanged connect_linux.sh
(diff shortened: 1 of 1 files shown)


outline/electron/connect_linux.sh
```diff
@@ -0,0 +1,21 @@
...
+# Instructions:
+# 1. Install the Outline client for Linux, connect to a server, and disconnect.
+#    This installs the outline controller service.
+# 2. $ git update-index --assume-unchanged connect_linux.sh
+#    This helps to avoid accidentally checking in your proxy credentials.
+# 3. Edit this script to add the IP, port, and password for your test proxy.
+# 4. $ ./connect_linux.sh
+# 5. Ctrl+C to stop proxying
+
+readonly PROXY_IP="..."
```
alalamav, line 13: It may be safer to assume these are exported variables or pass them as arguments to the script, preventing us from accidentally commiting the variables.

bemasc: Good idea!  I moved them to the command line.

alalamav (approved):

PR #94, opened 2022-10-21
DRAFT Add shadowsocks prefix support

bemasc:
This change is blocked pending the release of
outline-ss-server v1.3.6


bemasc:
@fortuna FYI.  I can update this once we have a new outline-ss-server release.

PR #96, opened 2022-10-24
cleanup: Refactor tunnel creation function arguments

bemasc:
This change moves Shadowsocks client setup to the caller, which avoids passing a large number of arguments across several packages and enables more flexibility for changes to the Shadowsocks client configuration (starting with #95).


fortuna (approved):
Awesome, thanks for the clean up!

PR #99, opened 2022-11-04
fix(apple, build): Add path extension to gomobile invocation

bemasc:
This is necessary for gomobile to find the gobind binary on my Mac.


fortuna (approved):
Interesting, that was not needed when I built the framework.

PR #118, opened 2023-04-14
fix: 💊 make sure gomobile can export CheckConnectivity and NewClient

jyyi1:
Outline Client (and other 3-rd party customers) is already referencing the following APIs (Android JNI interface) exported by `gomobile`. But due to the recent refactoring, we introduced types that are not recognized by `gomobile`.

image

In this PR, I re-exported these functions (together with the new `NewClientFromJSON`) by using the built-in types. Here is the exported Android JNI interface:

image

### DIFF
--- a/outline/electron/main.go
+++ b/outline/electron/main.go
@@ -103,7 +103,7 @@ func main() {
 		if err != nil {
 			log.Errorf("Failed to perform connectivity checks: %v", err)
 		}
-		os.Exit(connErrCode.Number())
+		os.Exit(connErrCode)
 	}
 
 	// Open TUN device
--- a/outline/shadowsocks/client.go
+++ b/outline/shadowsocks/client.go
@@ -28,7 +28,6 @@ import (
 	"github.com/Jigsaw-Code/outline-go-tun2socks/outline"
 	"github.com/Jigsaw-Code/outline-go-tun2socks/outline/connectivity"
 	"github.com/Jigsaw-Code/outline-go-tun2socks/outline/internal/utf8"
-	"github.com/Jigsaw-Code/outline-go-tun2socks/outline/neterrors"
 	"github.com/Jigsaw-Code/outline-internal-sdk/transport"
 	"github.com/Jigsaw-Code/outline-internal-sdk/transport/shadowsocks"
 	"github.com/Jigsaw-Code/outline-internal-sdk/transport/shadowsocks/client"
@@ -36,7 +35,12 @@ import (
 )
 
 // A client object that can be used to connect to a remote Shadowsocks proxy.
-type Client = outline.Client
+//
+// Somehow redundant because from it seems gomobile only recognizes a struct,
+// but not a type alias (i.e. type Client = outline.Client).
+type Client struct {
+	outline.Client
+}
 
 // NewClient creates a new Shadowsocks client from a non-nil configuration.
 //
@@ -98,7 +102,7 @@ func newShadowsocksClient(host string, port int, cipherName, password string, pr
 		return nil, fmt.Errorf("Failed to create PacketListener: %w", err)
 	}
 
-	return &outline.Client{StreamDialer: streamDialer, PacketListener: packetListener}, nil
+	return &Client{outline.Client{StreamDialer: streamDialer, PacketListener: packetListener}}, nil
 }
 
 const reachabilityTimeout = 10 * time.Second
@@ -107,8 +111,11 @@ const reachabilityTimeout = 10 * time.Second
 // the current network. Parallelizes the execution of TCP and UDP checks, selects the appropriate
 // error code to return accounting for transient network failures.
 // Returns an error if an unexpected error ocurrs.
-func CheckConnectivity(client *Client) (neterrors.Error, error) {
-	return connectivity.CheckConnectivity(client)
+//
+// Note: please make sure the return type is (int, error) for backward compatibility reason.
+func CheckConnectivity(client *Client) (int, error) {
+	netErr, err := connectivity.CheckConnectivity(&client.Client)
+	return netErr.Number(), err
 }
 
 // CheckServerReachable determines whether the server at `host:port` is reachable over TCP.


outline/shadowsocks/client.go
```diff
@@ -28,15 +28,19 @@ import (
...
 	"github.com/Jigsaw-Code/outline-internal-sdk/transport/shadowsocks/client"
 	"github.com/eycorsican/go-tun2socks/common/log"
 )
 
 // A client object that can be used to connect to a remote Shadowsocks proxy.
-type Client = outline.Client
+//
+// Somehow redundant because from it seems gomobile only recognizes a struct,
+// but not a type alias (i.e. type Client = outline.Client).
+type Client struct {
```
fortuna, line 41: Should we add the `outline` package to the go mobile call? That should fix this issue and it will be cleaner because it's decoupled from "shadowsocks".

jyyi1: We can export `outline` package, this will change the signature if we are defining type alias `type Client = outline.Client` (The `Client` class is changed from `shadowsocks.Client` to `outline.Client`, check the following JNI interface).

image

Anyway, we already introduced a breaking change in the signature (Now [`OutlineTunnel` is an alias of `Tunnel`](https://github.com/Jigsaw-Code/outline-go-tun2socks/blob/master/outline/tun2socks/tunnel.go#L43), so the generated JNI code will always use `Tunnel` instead of the old `OutlineTunnel`). We just need to decide which way is better:

* Redefining the exported package and update the client
* Creating duplicated types for backward compatibility

outline/connectivity.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package connectivity
+package outline
```
fortuna, line 15: I think this will break the old CheckConnectivity

outline/client.go
fortuna, line 1: I realized that we have a problem: we can't have the package `outline` be the only one exposed because that causes circular dependencies. The type definitions can't be declared on `outline` and implemented in other packages. Ugh, this is a pain, especially with the backward compatibility.

We may need to move things around

fortuna (changes requested):

outline/shadowsocks/legacy_api.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// ToOutlineClient retrieves the internal outline.Client from a legacy
+// shadowsocks.Client instance. This function is only for migrating from the
+// legacy API to the new outline API.
+//
+// Deprecated: use outline.Client directly.
+func ToOutlineClient(client *Client) *outline.Client {
+	return (*outline.Client)(client)
+}
```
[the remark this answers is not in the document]
jyyi1: Yes we are using it in electron. I was trying to hide the implementation detail here, but if we are OK with exposing it, I can remove this.

outline/client.go
jyyi1: Let me try interfaces (for example, define `outline.Client` as an interface instead of a `struct`). In that case, we don't need to reference any types defined in `outline`, just providing implementations in sub-packages. This will also resolve the above `NewShadowsocksConn` function cuz we can directly returns an internal `Client` implementation and name it as `NewShadowsocksClient`.

I can double confirm whether the generated JNI interface are identical to the old ones.

outline/connectivity.go
jyyi1: This will not break Outline Client, it is using `shadowsocks.CheckConnectivity`: [Android](https://github.com/Jigsaw-Code/outline-client/blob/c29197957a6cb2ea84ba4d619d97f4d836f98094/src/cordova/android/OutlineAndroidLib/outline/src/main/java/org/outline/vpn/VpnTunnelService.java#L370), [Apple](https://github.com/Jigsaw-Code/outline-client/blob/c29197957a6cb2ea84ba4d619d97f4d836f98094/src/cordova/apple/OutlineAppleLib/Sources/PacketTunnelProviderSources/PacketTunnelProvider.m#L141). But you are right, it might break other 3rd party users.

outline/internal/shadowsocks/shadowsocks.go
```diff
@@ -0,0 +1,81 @@
...
+	"fmt"
+	"net"
+
+	"github.com/Jigsaw-Code/outline-internal-sdk/transport"
+	"github.com/Jigsaw-Code/outline-internal-sdk/transport/shadowsocks"
+	"github.com/Jigsaw-Code/outline-internal-sdk/transport/shadowsocks/client"
+	"github.com/eycorsican/go-tun2socks/common/log"
+)
+
+func NewShadowsocksConn(host string, port int, cipherName, password string, prefix []byte) (*client.StreamDialer, *transport.PacketListener, error) {
```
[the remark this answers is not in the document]
jyyi1: Refactored `outline.Client` from a `struct` to `interface`, so we can implicitly implement it in sub-packages. Renamed this function to `NewShadowsocksClient`.

outline/client.go
```diff
@@ -25,3 +29,28 @@ type Client struct {
...
+			prefixBytes = p
+		}
+	}
+
+	sd, pl, err := internal.NewShadowsocksConn(config.Host, int(config.Port), config.Method, config.Password, prefixBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Client{*sd, *pl}, nil
```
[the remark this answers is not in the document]
jyyi1: It will introduce cyclic dependency in the old architecture. Also I would try to avoid any more dependencies to the deprecated `shadowsocks` package (rather I would depend on the `internal/shadowsocks` package), so in the future we can remove legacy packages.

outline/client.go
jyyi1: Refactored. For the new architecture, the sub-packages do not need to depend on the parent `outline` package any more, instead they can implicitly implement the interfaces defined in `outline` package (for example, `outline.Client` interface).

outline/tun2socks/tunnel_android.go
```diff
@@ -40,7 +40,7 @@ func init() {
 //
 // Returns an error if the TUN file descriptor cannot be opened, or if the tunnel fails to
 // connect.
-func ConnectShadowsocksTunnel(fd int, client *shadowsocks.Client, isUDPEnabled bool) (Tunnel, error) {
+func ConnectShadowsocksTunnel(fd int, client *shadowsocks.Client, isUDPEnabled bool) (OutlineTunnel, error) {
```
fortuna, line 43: Does this break backwards compatibility? If we are breaking it, then keeping the legacy api doesn't help much.

outline/tun2socks/tunnel_darwin.go
```diff
@@ -50,7 +50,7 @@ func init() {
 // `isUDPEnabled` indicates whether the tunnel and/or network enable UDP proxying.
 //
 // Sets an error if the tunnel fails to connect.
-func ConnectShadowsocksTunnel(tunWriter TunWriter, client *shadowsocks.Client, isUDPEnabled bool) (Tunnel, error) {
+func ConnectShadowsocksTunnel(tunWriter TunWriter, client *shadowsocks.Client, isUDPEnabled bool) (OutlineTunnel, error) {
```
fortuna, line 53: Same.

outline/client.go
```diff
@@ -15,13 +15,43 @@
...
+	internal "github.com/Jigsaw-Code/outline-go-tun2socks/outline/internal/shadowsocks"
+	"github.com/Jigsaw-Code/outline-go-tun2socks/outline/internal/utf8"
 	"github.com/Jigsaw-Code/outline-internal-sdk/transport"
 )
 
 // Client provides a transparent container for [transport.StreamDialer] and [transport.PacketListener]
 // that is exportable (as an opaque object) via gobind.
 // It's used by the connectivity test and the tun2socks handlers.
-type Client struct {
+type Client interface {
```
fortuna, line 28: If `Client` is to be a type used across packages, it perhaps needs its own package. `NewClientFromJSON` could be in another package.

fortuna (commented):
Are you in the office tomorrow? perhaps we can brainstorm a bit on the api

jyyi1:
> 

Will be there on Wednesday, let me schedule a meeting.

outline/client.go
jyyi1: `NewClientFromJSON` is not reused, but we need some `internal.NewShadowsocksClient` to be shared.

outline/tun2socks/tunnel_android.go
jyyi1: Changing from `Tunnel` to `OutlineTunnel` will keep backward compatibility, this is the old JNI interface (it requires `OutlineTunnel` instead of `Tunnel`):

image

jyyi1:
As discussed offline, @fortuna please take a look at the newly proposed `tun2socks.ConnectTunnel`, while still keeping all old function signature unchanged.

outline/tun2socks/tunnel.go
```diff
@@ -40,7 +44,7 @@ type Tunnel interface {
 }
 
 // Deprecated: use Tunnel directly.
-type OutlineTunnel = Tunnel
+type OutlineTunnel Tunnel
 
 type outlinetunnel struct {
 	tunnel.Tunnel
```
fortuna, line 50: Let's remove this. I don't really think it's needed, and it's extremely confusing. Let's see what breaks and fix it. I believe the comments here will address most issues.
```suggestion
```

outline/tun2socks/tunnel_android.go
```diff
@@ -52,3 +54,29 @@ func ConnectShadowsocksTunnel(fd int, client *shadowsocks.Client, isUDPEnabled b
...
+// ConnectTunnel reads packets from a TUN device represented by `fd` and routes
+// them to a remote proxy server represented by `configJSON`. This function
+// will also do connectivity tests before starting, so the caller is not
+// required to do any UDP tests.
+//
+// If the function succeeds, it will return a nil error and a Tunnel instance.
+//
+// This function does *not* take ownership of the TUN file descriptor `fd`; the
+// caller is responsible for closing after Tunnel disconnects.
+func ConnectTunnel(configJSON string, fd int) (Tunnel, error) {
```
fortuna, line 67: The ConnectTunnel function is doing different things in different platforms. That is confusing. We need to fix that.

Can you make this take a ReadWriteCloser for the tun device and move this to the shared code?

outline/tun2socks/tunnel_android.go
```diff
@@ -52,3 +54,29 @@ func ConnectShadowsocksTunnel(fd int, client *shadowsocks.Client, isUDPEnabled b
...
+	if len(configJSON) == 0 {
+		return nil, errors.New("tunnel configuration is required")
+	}
+	tunDev, err := tunnel.MakeTunFile(fd)
+	if err != nil {
+		return nil, err
+	}
+	tn, err := newTunnelFromJSON(configJSON, tunDev)
```
fortuna, line 75: Remove `tunDev`, and give a better name for the proxy tunnel:
```suggestion
	proxyTun, err := newTunnelFromJSON(configJSON)
```

The initialization that needs `tunDev` should happen below.

outline/tun2socks/tunnel_android.go
```diff
@@ -52,3 +54,29 @@ func ConnectShadowsocksTunnel(fd int, client *shadowsocks.Client, isUDPEnabled b
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to create tunnel from config: %w", err)
+	}
+
+	go tunnel.ProcessInputPackets(tn, tunDev)
```
fortuna, line 80: Use CopyBuffer. See comment on Electron code, which should be shared.

outline/tun2socks/tunnel_electron.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	if tunDev == nil {
+		return nil, errors.New("tun device is required")
+	}
+	tn, err := newTunnelFromJSON(configJSON, tunDev)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create tunnel from config: %w", err)
+	}
+
+	go io.CopyBuffer(tn, tunDev, make([]byte, mtu))
```
fortuna, line 47: We need to see the copy both ways in the same place. This code only shows one way, which is confusing. Where's the flow from the tun device to the proxy device? It takes a lot of digging to figure that out.

I think we could implement the bridge like this:
```suggestion
	go func() {
	  defer proxyTun.Close()
	  io.CopyBuffer(proxyDevice, tunDev, make([]byte, mtu))
	}
	go proxyTun.WriteTo(tunDev)
```

The reason for WriteTo is to avoid a layer of buffers when possible.

proxyTun.WriteTo will effectively call `core.RegisterOutputFn` to write to the tun device, Then wait for proxyTun.Close(). `ProxyTun.Close()` should call `lwipStack.Close()`. This way you don't need `tunnel.Tunnel` anymore.

fortuna (changes requested):

outline/tun2socks/tunnel_electron.go
```diff
@@ -0,0 +1,50 @@
...
+//
+// This function will close `tunDev` after Tunnel disconnects.
+func ConnectTunnel(configJSON string, tunDev io.ReadWriteCloser) (Tunnel, error) {
+	if len(configJSON) == 0 {
+		return nil, errors.New("tunnel configuration is required")
+	}
+	if tunDev == nil {
+		return nil, errors.New("tun device is required")
+	}
+	tn, err := newTunnelFromJSON(configJSON, tunDev)
```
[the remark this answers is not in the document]
jyyi1: I agree. `Tunnel` is a bad name, it's similar to tun device, which is very confusing.

outline/electron/main.go
```diff
@@ -91,14 +93,81 @@ func main() {
...
+			config.Prefix = p
+		}
+	}
+	client, err := shadowsocks.NewClient(&config)
 	if err != nil {
 		log.Errorf("Failed to create Shadowsocks client: %v", err)
 		os.Exit(neterrors.IllegalConfiguration.Number())
 	}
 
 	if *args.checkConnectivity {
```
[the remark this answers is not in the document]
jyyi1: We don't need any connectivity test for the new json config tunnel. I'd rather expose a public `OutlineTunnel.IsUDPSupported()` method so the caller can retrieve that info.

outline/internal/shadowsocks/shadowsocks.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, nil, fmt.Errorf("invalid shadowsocks configuration: %w", err)
+	}
+
+	// TODO: consider using net.LookupIP to get a list of IPs, and add logic for optimal selection.
+	proxyIP, err := net.ResolveIPAddr("ip", host)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to resolve proxy address: %w", err)
+	}
+
+	proxyTCPEndpoint := transport.TCPEndpoint{RemoteAddr: net.TCPAddr{IP: proxyIP.IP, Port: port}}
```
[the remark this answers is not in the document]
fortuna: FYI: https://github.com/Jigsaw-Code/outline-go-tun2socks/pull/120

Let me know of you want to merge that or not. It shouldn't conflict much.

